PR #3, opened 2020-06-02
Week 3: basic servlet implementation

hollyyuqizheng:
Links for Week 3 
[step 1](https://github.com/googleinterns/step/blob/master/walkthroughs/week-3-server/step-1-servlets-walkthrough.md)
[step 2](https://github.com/googleinterns/step/blob/master/walkthroughs/week-3-server/step-2-fetch-walkthrough.md)
[step 3](https://github.com/googleinterns/step/blob/master/walkthroughs/week-3-server/step-3-json-walkthrough.md)
[step 4](https://github.com/googleinterns/step/blob/master/walkthroughs/week-3-server/step-4-post-walkthrough.md)
[step 5](https://github.com/googleinterns/step/blob/master/walkthroughs/week-3-server/step-5-datastore-walkthrough.md)
[step 6](https://github.com/googleinterns/step/blob/master/walkthroughs/week-3-server/step-6-additional-features-walkthrough.md)

After following the first 3 steps of this week's walkthrough, I have added functions in the DataServlet class where I convert a few quotes into JSON format, which is then fetched by the getQuotes function in the script.js file. These quotes are then added to the page as a list of quotes. 

[Update 6.2] After completing step 4, I changed the quote section into a new section (in yellow). Now, the user can input a quote in the survey bar, and after the Submit button is clicked, the quote will appear below the survey bar. 

[Update 6.4] After completing steps 5 and 6, I updated the quote section with 1) refresh button, 2) a drop down menu for the user to set maximum number of quotes to be displayed 3) the GET and POST methods now utilize Datastore so that the quotes are provided by the user and stored in Datastore.  

Screenshot 2020-06-04 at 9 40 22 AM


portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,51 @@
...
+import com.google.gson.Gson;
 
-/** Servlet that returns some example content. TODO: modify this file to handle comments data */
+/** Servlet that returns some example content. */
 @WebServlet("/data")
 public class DataServlet extends HttpServlet {
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    ArrayList<String> quotes = createQuotesArray();
```
jamesnam-google: Here we can use List instead of ArrayList.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,51 @@
...
+
     response.setContentType("text/html;");
-    response.getWriter().println("<h1>Hello world!</h1>");
+    response.getWriter().println(quotesJson);
+  } 
+
+  /**
+   * This function creates an ArrayList containing a few hard-coded quotes. 
+   */ 
+  private ArrayList createQuotesArray() {
```
jamesnam-google: This method can return List. And can we specify the type?

ex:

private List<String> ...

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,51 @@
...
+    ArrayList<String> quotes = new ArrayList<String>();
```
jamesnam-google: And this line can be,

List<String> quotes = new ArrayList...

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
jamesnam-google: I think this method can be static.

private static ...

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,51 @@
...
+
+  /**
+   * This function takes in an ArrayList of quotes and converts it into a string 
+   * that is in JSON format. 
+   * Right now, this conversion is hard-coded. 
+   * Input type: ArrayList<String>
+   * Output type: String 
+   */ 
+  private String convertToJson(ArrayList<String> quotes){  
+    // String json = "{";
```
jamesnam-google: Did you want to delete these comments?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,51 @@
...
+    ArrayList<String> quotes = createQuotesArray();
+    String quotesJson = convertToJson(quotes);
+
     response.setContentType("text/html;");
-    response.getWriter().println("<h1>Hello world!</h1>");
+    response.getWriter().println(quotesJson);
+  } 
+
+  /**
+   * This function creates an ArrayList containing a few hard-coded quotes.
```
jamesnam-google: Consider following the javadoc pattern. Here and below.

https://en.wikipedia.org/wiki/Javadoc

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,51 @@
...
+  }
+
+  /**
+   * This function takes in an ArrayList of quotes and converts it into a string 
+   * that is in JSON format. 
+   * Right now, this conversion is hard-coded. 
+   * Input type: ArrayList<String>
+   * Output type: String 
+   */ 
+  private String convertToJson(ArrayList<String> quotes){
```
jamesnam-google: We need a space before {

portfolio/src/main/webapp/index.html
```diff
@@ -34,25 +34,26 @@ <h2>A few bullet points about me</h2>
...
+          <h2 onClick="getQuotes()">(Click me) Some of my favorite quotes from books</h2>
+          <ul id="quoteWrapper"></ul>
         </div> 
       
         <div id="projects" class="navbarSection">
           <h2>What I have been working on</h2> 
           <h3>(Click on each project to see more details!)</h3> 
           <div id="projectWrapper">
-            <div id="projectNlp" onClick="addProjectDescription('Nlp')"> 
+            <div id="projectNLP" onClick="addProjectDescription('NLP')">
```
jamesnam-google: I would recommend that we adhere to the camel case, even with acronyms.

ex:

projectNlp
projectDescriptionNlp
...

portfolio/src/main/webapp/index.html
```diff
@@ -34,25 +34,26 @@ <h2>A few bullet points about me</h2>
...
           <div id="projectWrapper">
-            <div id="projectNlp" onClick="addProjectDescription('Nlp')"> 
+            <div id="projectNLP" onClick="addProjectDescription('NLP')"> 
               <h3>Natural Language Processing research</h3>
             </div> 
-            <div id="projectDescriptionNlp"></div>
+            <div id="projectDescriptionNLP"></div>
 
-            <div id="projectTa" onClick="addProjectDescription('Ta')"> 
+            <div id="projectTA" onClick="addProjectDescription('TA')">
```
jamesnam-google: Same here and below.

walkthroughs/week-3-server/examples/server-stats/src/main/java/com/google/sps/servlets/ServerStatsServlet.java
```diff
@@ -15,7 +15,7 @@
 package com.google.sps.servlets;
 
 import com.google.sps.data.ServerStats;
-import com.google.gson.Gson;
+'import com.google.gson.Gson;'
```
jamesnam-google: Was this change intentional?

jamesnam-google (changes requested):

portfolio/src/main/webapp/script.js
```diff
@@ -92,4 +94,35 @@ function highlightSection(selectedIdString) {
...
+ * below the bulletin points on the page. 
+ */
+function getQuotes() {
+  fetch('/data')
+      .then(response => response.json())
+      .then((quotesJson) => {
+        const quoteWrapper = document.getElementById('quoteWrapper');
+        quoteWrapper.innerHTML = '';
+
+        Object.keys(quotesJson).forEach(
```
rgcase: Instead of looping over the keys, just to pull out the value, maybe we should just loop over the values?

rgcase (changes requested):

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done. I actually deleted this method in the new commit, because I have changed the quote creation to not be hard-coded any more.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Same as above.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Same as above.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/webapp/index.html
hollyyuqizheng: Done. Another change I made was changing the parameter passed into the addProjectDescription method : because I need the parameter to match the style of each element's naming, now I pass in "Nlp" instead of "NLP" as the parameter so that "projectDescriptionNlp" can be found. I think passing in "NLP" in all-caps and then trying to re-format this string into "Nlp" in the JavaScript file will be unnecessarily complicated.

portfolio/src/main/webapp/index.html
hollyyuqizheng: Done. Same as above.

walkthroughs/week-3-server/examples/server-stats/src/main/java/com/google/sps/servlets/ServerStatsServlet.java
hollyyuqizheng: It was not -- I think it was a typo on my end.

portfolio/src/main/webapp/script.js
hollyyuqizheng: Yes that would be better. I didn't realize that we can loop over the values directly.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,126 @@
...
+    * and puts them into a list of quotes. This list is then written
+    * as the response from the server. 
+    */
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Query query = new Query("Quote").addSort("timestamp", SortDirection.DESCENDING);
+    PreparedQuery quotes = datastore.prepare(query);
+
+    //Add each quote to the quote list whose content will be written as
```
jamesnam-google: Let's have a space after //. Here and below.

ex:

// Add each quote ...

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,126 @@
...
+  }
+
+  /**
+   * This method receives the quote that the user has inputed in the survey bar.
+   * This new quote is changed into an Entity that is put into Datastore.
+   * Each quote entity has a "text" and a "timstamp" field. 
+   */ 
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
```
jamesnam-google, line 87: nit: Unnecessary line break.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,126 @@
...
+    //the response from the servlet 
+    List<String> quoteList = new ArrayList<>();
+    int quoteCount = 1; 
+    for (Entity quoteEntity : quotes.asIterable()) {
+      if (quoteCount > numQuoteToDisplay) {
+        break; 
+      }
+      String quoteText = (String) quoteEntity.getProperty("text");
+      quoteList.add(quoteText);
+      quoteCount ++;
```
jamesnam-google: We don't need the space before ++

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,126 @@
...
+    quoteEntity.setProperty("timestamp", timestamp);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    datastore.put(quoteEntity);
+
+    response.sendRedirect("/index.html");
+  }
+
+  /**
+   * This method handles the returning of request parameter
```
jamesnam-google: Need period at the end of the sentence.

portfolio/src/main/webapp/WEB-INF/appengine-web.xml
```diff
@@ -1,12 +1,10 @@
...
-  <version>1</version>
   <threadsafe>false</threadsafe>
   <sessions-enabled>true</sessions-enabled>
   <runtime>java8</runtime>
   <static-files>
     <!-- prevent unwanted caching when accessing via the web preview server -->
     <include path="/**" expiration="0s" />
   </static-files>
-</appengine-web-app>
+</appengine-web-app>
```
jamesnam-google: Let's try to have newline at the end of files.

Reference:
https://stackoverflow.com/questions/5813311/no-newline-at-end-of-file

Here and below.

portfolio/src/main/webapp/style.css
```diff
@@ -48,6 +54,7 @@
   margin-left: 10%;
 }
 
+
```
jamesnam-google: Unnecessary line break

portfolio/src/main/webapp/script.js
```diff
@@ -57,39 +63,83 @@ function addProjectDescription(project) {
...
+
+/**
+ * This function performs a POST request to deleteQuote servlet, which 
+ * handles the deletion of all quotes in Datastore. 
+ */
+function cleanData() {
+  fetch('/deleteQuote', {method: 'POST'})
+    .then(() => {
+      refreshQuotes(); 
+  });
```
jamesnam-google: I think we need to indent this line by two spaces.

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/webapp/WEB-INF/appengine-web.xml
hollyyuqizheng: Done.

portfolio/src/main/webapp/style.css
hollyyuqizheng: Done.

portfolio/src/main/webapp/script.js
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,129 @@
...
+import com.google.gson.Gson;
+import java.util.List; 
+import java.util.ArrayList; 
 
-/** Servlet that returns some example content. TODO: modify this file to handle comments data */
+/** Servlet that returns some example content. */
 @WebServlet("/data")
 public class DataServlet extends HttpServlet {
 
+  int numQuoteToDisplay = 10;
```
rgcase: This seems like it's meant to be a constant value within the class? If so can we make it:
```
private static final int NUM_QUOTES_TO_DISPLAY = 10;
```

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,129 @@
...
+
+  /**
+    * This method retrives all quotes that are stored in Datastore
+    * and puts them into a list of quotes. This list is then written
+    * as the response from the server. 
+    */
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Query query = new Query("Quote").addSort("timestamp", SortDirection.DESCENDING);
```
rgcase: It looks like we use these Strings "Quote" and "timestamp" in a couple other places too. Can you extract them out at `private static final` constants at the top of this class so we don't accidentally change one without changing all of them?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,129 @@
...
+    PreparedQuery quotes = datastore.prepare(query);
+
+    // Add each quote to the quote list whose content will be written as 
+    // the response from the servlet. 
+    List<String> quoteList = new ArrayList<>();
+    int quoteCount = 1; 
+    for (Entity quoteEntity : quotes.asIterable()) {
```
rgcase: I know that the instructions told you to use `quotes.asIterable()` but I think it would be cleaner and more efficient to set a limit on the query itself, rather than doing it after the fact as you are here. You can look in the API for PreparedQuery https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/PreparedQuery#interface-preparedquery.

I think it'll look something like:
```
quotes.asList(FetchOptions.Builder.withLimit(NUM_QUOTES_TO_DISPLAY)).forEach(quote -> { ... });
```

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,123 @@
...
+      String text = (String) quoteEntity.getProperty("text");
+      long timestamp = (long) quoteEntity.getProperty("timestamp");
+
+      Quote newQuote = new Quote(id, text, timestamp);
+      quoteList.add(newQuote);
+      quoteCount++; 
+    }
+
+    String quotesJson = convertToJson(quoteList);
     response.setContentType("text/html;");
```
rgcase: I think we need to update this context type to "application/json"

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,123 @@
...
+import java.util.List; 
+import java.util.ArrayList; 
 
-/** Servlet that returns some example content. TODO: modify this file to handle comments data */
+/** Servlet that returns some example content. */
 @WebServlet("/data")
 public class DataServlet extends HttpServlet {
 
+  /**
+    * This method retrives all quotes that are stored in Datastore
```
rgcase: I think you can remove the "This method..." or "This function..." from your javadoc comments because it's not really adding anything valuable. This one can just start with "Retrieves all quotes..." and similar for the others.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,123 @@
...
+
+    // Add each quote to the quote list whose content will be written as 
+    // the response from the servlet. 
+    List<Quote> quoteList = new ArrayList<>();
+    int quoteCount = 1; 
+    for (Entity quoteEntity : quotes.asIterable()) {
+      if (quoteCount > numQuoteToDisplay) {
+        break; 
+      }
+      long id = quoteEntity.getKey().getId();
```
rgcase: It would be nice to use a helper method so you could just do
```
Quote quote = extractQuote(quoteEntity);
```

because the details of converting an `Entity` to a `Quote` are not really important to the logic in this method, and later you can reuse it in other places as well.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,123 @@
...
+    String json = gson.toJson(quoteList);
+    return json;
+  }
+
+  /**
+   * This method reads the numToDisplay parameter that is set by the user from the survey bar. 
+   * @param request: the request; defaultNum: default number to display set in doGet method. 
+   * @return number of quotes to display 
+   */ 
+  private int getNumQuoteDispalyed(HttpServletRequest request, int defaultNumToDisplay) {
```
rgcase: nit: Typo Dispalyed -> Displayed

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,123 @@
...
+    int numDisplay;
```
rgcase: I generally try not to follow this pattern of declaring a variable and then updating it later based on some condition. It can sometimes make your code hard to follow because you have always know all the places that may have updated the variable to know the value. In fact it's usually better if you can to implicitly assume your variables are `final` and to not mutate them if possible.

In this case I don't think you actually even need the variable, and you can just do:

```
try {
  return Integer.parseInt(numQuoteToDisplayParam);
} catch (NumberFormatException e) {
  return defaultNumToDisplay;
}
```

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,123 @@
...
+    * as the response from the server. 
+    */
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Query query = new Query("Quote").addSort("timestamp", SortDirection.DESCENDING);
+    PreparedQuery quotes = datastore.prepare(query);
+
+    // Get the parameter for number to display, with 5 as the default value.
+    int numQuoteToDisplay = getNumQuoteDispalyed(request, 5);
```
rgcase: Rather than having this stray number 5 here, I suggest to extract it out into a constant at the top of this class like:

```
private static final int DEFAULT_NUM_QUOTES = 5;
```

and then you can reference that directly inside `getNumQuoteDisplayed` rather than pass it in, unless you plan to have multiple different defaults.

portfolio/src/main/java/com/google/sps/servlets/DeleteDataServlet.java
```diff
@@ -0,0 +1,45 @@
...
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Query query = new Query("Quote");
+    PreparedQuery quotes = datastore.prepare(query);
+
+    for (Entity quoteEntity : quotes.asIterable()) {
+      Key quoteEntityKey = quoteEntity.getKey();
+      datastore.delete(quoteEntityKey);
```
rgcase: It's not that important, but you can also pass a `List<Key>` to the `datastore.delete(...)` method and then they'll be deleted in a single call to the datastore service instead of individual calls for each entity which should be much more efficient. Maybe build a `List` of the keys and then pass them all at the same time?

rgcase (changes requested):

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: I see -- yes, this variable is better as a static final variable.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done. Same as above.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done. It's beneficial that this refactoring avoids using the counter and the break statement.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DeleteDataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
rgcase: It looks like you used them everywhere except in this line, can you use the constants here too?

rgcase (approved):

portfolio/pom.xml
```diff
@@ -1,4 +1,4 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0"
+ <project xmlns="http://maven.apache.org/POM/4.0.0"
```
jamesnam-google, line 1: Unnecessary change

portfolio/src/main/java/com/google/sps/data/Quote.java
```diff
@@ -0,0 +1,15 @@
+package com.google.sps.data;
+
+/** Class that models a quote. */
+public final class Quote {
+
+  private final long id;
+  private final String text;
+  private final long timestamp;
```
jamesnam-google: What is the unit of this timestamp? Consider updating the variable name, here and elsewhere.

Ex:
timestampMillis

portfolio/src/main/webapp/script.js
```diff
@@ -57,39 +63,90 @@ function addProjectDescription(project) {
...
+/**
+ * This function fetches the text from /data, with additional parameters
+ * added to the request's URL string. 
+ * Then, the function puts the received text into the quoteWrapper
+ * that gets displayed on the UI.  
+ */
+function getQuotes() {
+  const mainURL = '/data';
+  const numToDisplayURL = '?numToDisplay=';
+  const numToDisplay = document.getElementById("numToDisplay").value;
```
jamesnam-google: Let's use single quotes for string literals across this file.

Reference:
https://google.github.io/styleguide/jsguide.html#features-strings-use-single-quotes

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,131 @@
...
+    PreparedQuery quotes = datastore.prepare(query);
+
+    // Get the parameter for number to display, with 5 as the default value.
+    int numQuoteToDisplay = getNumQuoteDisplayed(request, DEFAULT_NUM_QUOTES);
+   
+    // Add each quote to the quote list whose content will be written as 
+    // the response from the servlet. 
+    List<Quote> quoteList = new ArrayList<>();
+    quotes.asList(FetchOptions.Builder.withLimit(numQuoteToDisplay))
+          .forEach((quoteEntity) -> {
```
jamesnam-google: Indentation seems too much here. I think we need four spaces from column aligned with q (in quotes)

portfolio/src/main/java/com/google/sps/servlets/DeleteDataServlet.java
```diff
@@ -0,0 +1,53 @@
...
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Query query = new Query(QUOTE);
+    PreparedQuery quotes = datastore.prepare(query);
+
+    List<Key> keyList = new ArrayList<Key>();
+    quotes.asIterable()
+          .forEach((quoteEntity) -> {
```
jamesnam-google: Similarly, indentation seems excessive.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,131 @@
...
+   * This new quote is changed into an Entity that is put into Datastore.
+   * Each quote entity has a "text" and a "timstamp" field. 
+   */ 
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {    
+    String quote = request.getParameter("quote");
+
+    // Only update the new quote if it is not an empty string. 
+    if (quote.length() > 0) {
+      long timestamp = System.currentTimeMillis();
```
jamesnam-google: Let's indicate the time unit in this variable as well.

ex:
timestampMillis

jamesnam-google (changes requested):

portfolio/pom.xml
hollyyuqizheng: I think there was an extra blank space or line break at the beginning of the file, so I think in this change I just deleted that.

portfolio/src/main/java/com/google/sps/data/Quote.java
hollyyuqizheng: Done.

portfolio/src/main/webapp/script.js
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DeleteDataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/pom.xml
jamesnam-google: Can we remove the unnecessary space in the beginning of this line?

portfolio/src/main/webapp/script.js
```diff
@@ -57,39 +63,90 @@ function addProjectDescription(project) {
...
       if (section.id !== selectedIdString) {
         section.style.border = 'none'; 
       }
     }); 
-}
\ No newline at end of file
+}
+
+/**
+ * This function fetches the text from /data, with additional parameters
```
jamesnam-google, line 99: We don't need "This function". We can just start the comment with

Fetches the text ...

portfolio/src/main/webapp/script.js
```diff
@@ -57,39 +63,90 @@ function addProjectDescription(project) {
...
+
+      Object.values(quotesJson).forEach(
+        (quote) => {
+          quoteListElement.appendChild(createQuotesListElement(quote)); 
+        });        
+    });
+}
+
+/**
+ * This is a helper function, called inside getQuotes.
```
jamesnam-google, line 124: Same here

portfolio/src/main/webapp/script.js
```diff
@@ -57,39 +63,90 @@ function addProjectDescription(project) {
...
+ * this li element will be added to the Quotes section in the HTML file.  
+ */
+function createQuotesListElement(quote) {
+  const liElement = document.createElement('li');
+  liElement.innerText = quote.text;
+  return liElement;
+}
+
+/**
+ * This function performs a POST request to deleteQuote servlet, which
```
jamesnam-google, line 135: Same here

portfolio/src/main/webapp/script.js
```diff
@@ -57,39 +63,90 @@ function addProjectDescription(project) {
...
+ */
+function cleanData() {
+  fetch('/deleteQuote', {method: 'POST'})
+    .then(() => {
+      refreshQuotes(); 
+    });
+}
+
+/**
+ * This function handles the visual refreshing of the quote list element
```
jamesnam-google, line 146: Same here

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -19,14 +19,139 @@
...
+    List<Quote> quoteList = new ArrayList<>();
+    quotes.asList(FetchOptions.Builder.withLimit(numQuoteToDisplay))
+        .forEach((quoteEntity) -> {
+          quoteList.add(extractQuoteFromEntity(quoteEntity));
+        });
+    
+    // Put the thread to sleep for a short period of time.
+    // This adds buffer time for querying Datastore. 
+    try {
+      Thread.sleep(200);
```
jamesnam-google, line 69: Do we absolutely need this? What happens if we remove the thread sleep? Please clarify these areas in the comment.

jamesnam-google (approved):
Approving the changes with some comments.

portfolio/src/main/webapp/script.js
hollyyuqizheng: Done.

portfolio/src/main/webapp/script.js
hollyyuqizheng: Done.

portfolio/src/main/webapp/script.js
hollyyuqizheng: Done.

portfolio/src/main/webapp/script.js
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: I will add more comments regarding this sleep function.

PR #4, opened 2020-06-08
Week 4: Login feature

hollyyuqizheng:
Login feature from Week 4: [link](https://github.com/googleinterns/step/blob/master/walkthroughs/week-4-libraries/authentication/authentication-walkthrough.md)

- If user is not logged in, the quote submission box is hidden, and the text on the right will direct the user to another page to log in
- Once logged in, the quote submission box appears. If the user doesn’t have a nickname yet, the textbox on the right will direct the user to another page to create a nickname.
--- If the user submits quotes at this time, the quote will be displayed with the author being “unknown user” 
- Once the user is logged in and has a nickname, the create nickname box will be hidden, and the log in status will display “Logged in as <nickname>”
--- If the user submits quotes at this time, the quote will be displayed with the nickname being the author associated with each quote. 
- If a user already with a nickname logs out and logs back in later, the login status will display “Logged in as <nickname>” instead of their email address.

### DIFF
--- a/portfolio/src/main/java/com/google/sps/data/Util.java
+++ b/portfolio/src/main/java/com/google/sps/data/Util.java
@@ -14,6 +14,10 @@ import java.util.Optional;
 
 public final class Util {
 
+  private static final String PROPERTY_NAME_USER_ID = "userId"; 
+  private static final String PROPERTY_NAME_USER_INFO = "UserInfo"; 
+  private static final String PROPERTY_NAME_NICKNAME = "nickname"; 
+
   /**
    * Sets up strong consistency for datastore service. 
    * This strong consistency ensures that freshness is more important than availability
@@ -31,14 +35,14 @@ public final class Util {
    */
   public Optional<String> getUserNickname(DatastoreService datastore, String userId) {
     Query query =
-        new Query("UserInfo")
-            .setFilter(new Query.FilterPredicate("userId", Query.FilterOperator.EQUAL, userId));
+        new Query(PROPERTY_NAME_USER_INFO)
+            .setFilter(new Query.FilterPredicate(PROPERTY_NAME_USER_ID, Query.FilterOperator.EQUAL, userId));
     PreparedQuery results = datastore.prepare(query);
     Entity entity = results.asSingleEntity();
 
     String nickname = null;
     if (entity != null) {
-      nickname = (String) entity.getProperty("nickname");
+      nickname = (String) entity.getProperty(PROPERTY_NAME_NICKNAME);
     }
 
     Optional<String> nicknameOptional = Optional.ofNullable(nickname);
--- a/portfolio/src/main/java/com/google/sps/data/Util.java
+++ b/portfolio/src/main/java/com/google/sps/data/Util.java
@@ -36,12 +36,10 @@ public final class Util {
     PreparedQuery results = datastore.prepare(query);
     Entity entity = results.asSingleEntity();
 
-    String nickname;
-    if (entity == null) {
-      nickname = null; 
-    } else {
+    String nickname = null;
+    if (entity != null) {
       nickname = (String) entity.getProperty("nickname");
-    } 
+    }
 
     Optional<String> nicknameOptional = Optional.ofNullable(nickname);
     return nicknameOptional;
--- a/portfolio/src/main/java/com/google/sps/data/Util.java
+++ b/portfolio/src/main/java/com/google/sps/data/Util.java
@@ -1,0 +1,49 @@
+package com.google.sps.data;
+
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.DatastoreServiceConfig;
+import com.google.appengine.api.datastore.DatastoreServiceConfig.Builder;  
+import com.google.appengine.api.datastore.ReadPolicy;
+import com.google.appengine.api.datastore.ReadPolicy.Consistency;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Query.SortDirection;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Entity;
+import java.util.Optional; 
+
+public final class Util {
+
+  /**
+   * Sets up strong consistency for datastore service. 
+   * This strong consistency ensures that freshness is more important than availability
+   * so that the most up-to-date data is returned and displayed on the page.
+   */ 
+  public static DatastoreService getDatastoreServiceWithConsistency() { 
+    DatastoreServiceConfig datastoreConfig = 
+        DatastoreServiceConfig.Builder.withReadPolicy(new ReadPolicy(Consistency.STRONG)).deadline(5.0);
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService(datastoreConfig);
+    return datastore; 
+  }
+
+  /**
+   * @return an Optional of type String that represents the user's nickname. 
+   */
+  public Optional<String> getUserNickname(DatastoreService datastore, String userId) {
+    Query query =
+        new Query("UserInfo")
+            .setFilter(new Query.FilterPredicate("userId", Query.FilterOperator.EQUAL, userId));
+    PreparedQuery results = datastore.prepare(query);
+    Entity entity = results.asSingleEntity();
+
+    String nickname;
+    if (entity == null) {
+      nickname = null; 
+    } else {
+      nickname = (String) entity.getProperty("nickname");
+    } 
+
+    Optional<String> nicknameOptional = Optional.ofNullable(nickname);
+    return nicknameOptional;
+  }
+}
--- a/portfolio/src/main/java/com/google/sps/data/Util.java
+++ b/portfolio/src/main/java/com/google/sps/data/Util.java
@@ -33,19 +33,18 @@ public final class Util {
   /**
    * @return an Optional of type String that represents the user's nickname. 
    */
-  public Optional<String> getUserNickname(DatastoreService datastore, String userId) {
+  public static Optional<String> getUserNickname(DatastoreService datastore, String userId) {
     Query query =
         new Query(PROPERTY_NAME_USER_INFO)
             .setFilter(new Query.FilterPredicate(PROPERTY_NAME_USER_ID, Query.FilterOperator.EQUAL, userId));
     PreparedQuery results = datastore.prepare(query);
     Entity entity = results.asSingleEntity();
-
     String nickname = null;
     if (entity != null) {
       nickname = (String) entity.getProperty(PROPERTY_NAME_NICKNAME);
     }
-
     Optional<String> nicknameOptional = Optional.ofNullable(nickname);
+
     return nicknameOptional;
   }
 }
--- a/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -26,7 +26,9 @@ import java.util.Optional;
 @WebServlet("/data")
 public class DataServlet extends HttpServlet {
 
+  // Constants for URL links. 
   private static final String INDEX_URL = "/index.html"; 
+  private static final String HOME_URL = "/"; 
 
   // Constants for property of a Quote item in Datastore.
   private static final String PROPERTY_NAME_QUOTE = "Quote";
@@ -36,22 +38,17 @@ public class DataServlet extends HttpServlet {
   private static final String PROPERTY_NAME_NICKNAME = "nickname";
 
   // Constants for information to be put into response of a GET request.
-  private static final String IS_USER_LOGGED_IN = "loggedIn"; 
+  private static final String IS_USER_LOGGED_IN_PARAM = "loggedIn"; 
   private static final String REDIRECT_URL_PARAM = "redirectUrl";
-  private static final String HOME_URL = "/"; 
-
+  
   // Constant for fetching quotes from Datastore
   private static final int DEFAULT_NUM_QUOTES = 5;
 
-  // Global variable for an instance of the Util helper class
-  private static Util util; 
-  
   // Global variable for an instance of DatastoreService
   private static DatastoreService datastore; 
 
   public DataServlet() {
-    util = new Util();
-    datastore = util.getDatastoreServiceWithConsistency(); 
+    datastore = Util.getDatastoreServiceWithConsistency(); 
   }
   
   @Override
@@ -66,10 +63,10 @@ public class DataServlet extends HttpServlet {
     if (userService.isUserLoggedIn()) {
       String logoutUrl = userService.createLogoutURL(HOME_URL);
 
-      Optional<String> nicknameOptional = util.getUserNickname(datastore, currentUser.getUserId());
+      Optional<String> nicknameOptional = Util.getUserNickname(datastore, currentUser.getUserId());
       nicknameOptional.ifPresent(nickname -> responseMap.put(PROPERTY_NAME_NICKNAME, nickname));
       
-      responseMap.put(IS_USER_LOGGED_IN, "true");
+      responseMap.put(IS_USER_LOGGED_IN_PARAM, "true");
       responseMap.put(REDIRECT_URL_PARAM, logoutUrl); 
 
       // Only display quotes if the user is logged in
@@ -78,7 +75,7 @@ public class DataServlet extends HttpServlet {
       responseMap.put(PROPERTY_NAME_QUOTE, quotesJson);
     } else {
       String loginUrl = userService.createLoginURL(HOME_URL);
-      responseMap.put(IS_USER_LOGGED_IN, "false");
+      responseMap.put(IS_USER_LOGGED_IN_PARAM, "false");
       responseMap.put(REDIRECT_URL_PARAM, loginUrl); 
     }
 
@@ -112,9 +109,7 @@ public class DataServlet extends HttpServlet {
   }
 
   /**
-   * Creates a new entity for each quote and puts each new quote into Datastore
-   * @param request: the HTTP request; 
-   *        userInfo: a HashMap that contains user information connected to each quote
+   * Creates a new entity for each quote and puts each new quote into Datastore. 
    */ 
   private void putQuoteIntoDatastore(HttpServletRequest request, User user) {
     String quote = request.getParameter("quote");   
@@ -126,7 +121,7 @@ public class DataServlet extends HttpServlet {
       quoteEntity.setProperty(PROPERTY_NAME_TIMESTAMP, timestampMillis);
       quoteEntity.setProperty(PROPERTY_NAME_USER_EMAIL, user.getEmail()); 
 
-      Optional<String> nicknameOptional = util.getUserNickname(datastore, user.getUserId());
+      Optional<String> nicknameOptional = Util.getUserNickname(datastore, user.getUserId());
       nicknameOptional.ifPresent(nickname -> quoteEntity.setProperty(PROPERTY_NAME_NICKNAME, nickname));
  
       datastore.put(quoteEntity);
--- a/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -183,7 +183,7 @@ public class DataServlet extends HttpServlet {
    * Returns the nickname of the user with id, or empty String if the user has not set a nickname.
    */
   private String getUserNickname(String id) {
-    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    DatastoreService datastore = getDatastoreServiceWithConsistency();
     Query query =
         new Query("UserInfo")
             .setFilter(new Query.FilterPredicate("id", Query.FilterOperator.EQUAL, id));
--- a/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -1,5 +1,6 @@
 package com.google.sps.servlets;
 
+import com.google.appengine.api.users.User; 
 import com.google.appengine.api.users.UserService;
 import com.google.appengine.api.users.UserServiceFactory;
 import com.google.appengine.api.datastore.DatastoreService;
@@ -21,7 +22,8 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import com.google.gson.Gson;
 import java.util.List; 
-import java.util.ArrayList; 
+import java.util.ArrayList;
+import java.util.Map; 
 import java.util.HashMap; 
 
 @WebServlet("/data")
@@ -30,41 +32,37 @@ public class DataServlet extends HttpServlet {
   private static final String QUOTE = "Quote";
   private static final String TEXT = "text";
   private static final String TIMESTAMP = "timestamp"; 
-  private static final String USEREMAIL = "userEmail";
-  private static final String LOGGEDIN = "loggedIn"; 
-  private static final String REDIRECTURL = "redirectUrl";
+  private static final String USER_EMAIL = "userEmail";
+  private static final String LOGGED_IN = "loggedIn"; 
+  private static final String REDIRECT_URL = "redirectUrl";
   private static final String NICKNAME = "nickname";
   private static final int DEFAULT_NUM_QUOTES = 5;
+  private static final String HOME_URL = "/"; 
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     response.setContentType("application/json");
 
     UserService userService = UserServiceFactory.getUserService();
-    HashMap<String, String> responseMap = new HashMap<String, String>();
+
+    Map<String, String> responseMap = new HashMap<String, String>();
+    User currentUser = userService.getCurrentUser(); 
 
     if (userService.isUserLoggedIn()) {
-      String userEmail = userService.getCurrentUser().getEmail();
-      String urlToRedirectToForLogOut = "/";
-      String logoutUrl = userService.createLogoutURL(urlToRedirectToForLogOut);
-      String nickname = getUserNickname(userService.getCurrentUser().getUserId());  
+      String logoutUrl = userService.createLogoutURL(HOME_URL);
+      String nickname = getUserNickname(currentUser.getUserId());  
 
-      responseMap.put(LOGGEDIN, "true");
-      responseMap.put(USEREMAIL, userEmail);
-      responseMap.put(REDIRECTURL, logoutUrl); 
+      responseMap.put(LOGGED_IN, "true");
+      responseMap.put(REDIRECT_URL, logoutUrl); 
       responseMap.put(NICKNAME, nickname); 
 
       // Only display quotes if the user is logged in
       String quotesJson = getQuoteListJson(request);
       responseMap.put(QUOTE, quotesJson);
     } else {
-      String urlToRedirectToForLogIn = "/";
-      String loginUrl = userService.createLoginURL(urlToRedirectToForLogIn);
-      responseMap.put(LOGGEDIN, "false");
-      responseMap.put(USEREMAIL, null);
-      responseMap.put(REDIRECTURL, loginUrl); 
-      responseMap.put(NICKNAME, null); 
-      responseMap.put(QUOTE, null);
+      String loginUrl = userService.createLoginURL(HOME_URL);
+      responseMap.put(LOGGED_IN, "false");
+      responseMap.put(REDIRECT_URL, loginUrl); 
     }
 
     Gson gson = new Gson();
@@ -77,21 +75,26 @@ public class DataServlet extends HttpServlet {
     UserService userService = UserServiceFactory.getUserService();
 
     if (userService.isUserLoggedIn()) {
-      String userEmail = userService.getCurrentUser().getEmail();
-      String nickname = getUserNickname(userService.getCurrentUser().getUserId()); 
-
-      // HashMap that contains user information: email address, nickname.
-      HashMap<String, String> userInfo = new HashMap<String, String>();
-      userInfo.put(USEREMAIL, userEmail);
-      userInfo.put(NICKNAME, nickname); 
-
-      putQuoteIntoDatastore(request, userInfo);
+      User currentUser = userService.getCurrentUser(); 
+      putQuoteIntoDatastore(request, currentUser);
     } 
 
     response.sendRedirect("/index.html");
   } 
 
-   /**
+  /**
+   * Sets up strong consistency for datastore service. 
+   * This Strong Consistency ensures that freshness is more important than availability
+   * so that the most up-to-date data is returned and displayed on the page.
+   */ 
+  private DatastoreService getDatastoreServiceWithConsistency() { 
+    DatastoreServiceConfig datastoreConfig = 
+        DatastoreServiceConfig.Builder.withReadPolicy(new ReadPolicy(Consistency.STRONG)).deadline(5.0);
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService(datastoreConfig);
+    return datastore; 
+  }
+
+  /**
    * Takes in an ArrayList of quotes and converts it into a string 
    * that is in JSON format. 
    * @param a list of quotes
@@ -108,7 +111,7 @@ public class DataServlet extends HttpServlet {
    * @param request: the HTTP request; 
    *        userInfo: a HashMap that contains user information connected to each quote
    */ 
-  private void putQuoteIntoDatastore(HttpServletRequest request, HashMap<String, String> userInfo) {
+  private void putQuoteIntoDatastore(HttpServletRequest request, User user) {
     String quote = request.getParameter("quote");   
     if (quote.length() > 0) {
       long timestampMillis = System.currentTimeMillis();
@@ -116,15 +119,10 @@ public class DataServlet extends HttpServlet {
       Entity quoteEntity = new Entity(QUOTE);
       quoteEntity.setProperty(TEXT, quote);
       quoteEntity.setProperty(TIMESTAMP, timestampMillis);
-      quoteEntity.setProperty(USEREMAIL, userInfo.get(USEREMAIL)); 
-      quoteEntity.setProperty(NICKNAME, userInfo.get(NICKNAME));
+      quoteEntity.setProperty(USER_EMAIL, user.getEmail()); 
+      quoteEntity.setProperty(NICKNAME, getUserNickname(user.getUserId()));
 
-      // This Strong Consistency ensures that freshness is more important than availability
-      // so that the most up-to-date data is returned and displayed on the page. 
-      DatastoreServiceConfig datastoreConfig = 
-          DatastoreServiceConfig.Builder.withReadPolicy(new ReadPolicy(Consistency.STRONG)).deadline(5.0);
-
-      DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+      DatastoreService datastore = getDatastoreServiceWithConsistency();
       datastore.put(quoteEntity);
     }
   }
@@ -144,9 +142,9 @@ public class DataServlet extends HttpServlet {
     // the response from the servlet. 
     List<Quote> quoteList = new ArrayList<>();
     quotes.asList(FetchOptions.Builder.withLimit(numQuoteToDisplay))
-          .forEach((quoteEntity) -> {
-            quoteList.add(extractQuoteFromEntity(quoteEntity));
-          });
+        .forEach((quoteEntity) -> {
+          quoteList.add(extractQuoteFromEntity(quoteEntity));
+        });
 
     String quotesJson = convertQuotesToJson(quoteList);
     return quotesJson; 
@@ -175,7 +173,7 @@ public class DataServlet extends HttpServlet {
     long id = quoteEntity.getKey().getId();
     String text = (String) quoteEntity.getProperty(TEXT);
     long timestampMillis = (long) quoteEntity.getProperty(TIMESTAMP);
-    String userEmail = (String) quoteEntity.getProperty(USEREMAIL); 
+    String userEmail = (String) quoteEntity.getProperty(USER_EMAIL); 
     String nickname = (String) quoteEntity.getProperty(NICKNAME); 
     Quote newQuote = new Quote(id, text, timestampMillis, userEmail, nickname);
     return newQuote; 
--- a/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -111,11 +111,11 @@ public class DataServlet extends HttpServlet {
   private void putQuoteIntoDatastore(HttpServletRequest request, HashMap<String, String> userInfo) {
     String quote = request.getParameter("quote");   
     if (quote.length() > 0) {
-      long timestamp = System.currentTimeMillis();
+      long timestampMillis = System.currentTimeMillis();
 
       Entity quoteEntity = new Entity(QUOTE);
       quoteEntity.setProperty(TEXT, quote);
-      quoteEntity.setProperty(TIMESTAMP, timestamp);
+      quoteEntity.setProperty(TIMESTAMP, timestampMillis);
       quoteEntity.setProperty(USEREMAIL, userInfo.get(USEREMAIL)); 
       quoteEntity.setProperty(NICKNAME, userInfo.get(NICKNAME));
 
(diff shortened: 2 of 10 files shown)


portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,143 +1,155 @@
...
+import java.util.HashMap; 
 
-/** Servlet that returns some example content. */
 @WebServlet("/data")
 public class DataServlet extends HttpServlet {
 
   private static final String QUOTE = "Quote";
   private static final String TEXT = "text";
   private static final String TIMESTAMP = "timestamp"; 
+  private static final String USEREMAIL = "userEmail";
```
jamesnam-google: Let's use underscore between terms.

ex:
USER_EMAIL
LOGGED_IN
REDIRECT_URL

But NICKNAME sounds good.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,143 +1,155 @@
...
-        });
-    
-    // Put the thread to sleep for a short period of time.
-    // This adds buffer time for querying Datastore. 
-    try {
-      Thread.sleep(200);
-    } catch (Exception e) {
-      System.out.println(e);
+    UserService userService = UserServiceFactory.getUserService();
+    HashMap<String, String> responseMap = new HashMap<String, String>();
```
jamesnam-google: We can use the interface as the type.

Map<...> ...

portfolio/src/main/webapp/script.js
```diff
@@ -143,10 +202,16 @@ function cleanData() {
...
  * by setting its content to empty. 
  */
 function refreshQuotes() {
   const quoteListElement = document.getElementById('quoteWrapper');
   quoteListElement.innerHTML = '';
 }
+
+/** Creates a map and adds it to the page. */
+function createMap() {
+  const map = new google.maps.Map(document.getElementById('simpleMap'),
```
jamesnam-google: I think it can be clearer to put the line break before document.

ex:

const map = new google.maps.Map(
  document...,
  {...});

jamesnam-google: Somehow GitHub comment does not show the indentation correctly. Please indent according to the style guide.

portfolio/src/main/webapp/script.js
```diff
@@ -109,30 +108,90 @@ function getQuotes() {
...
+/**
+ * Displays the username's nickname if a nickname is stored in the response
+ * Json returned by GET request. If a nickname does not exist,
+ * the textbox prompting the user to create a nickname appears. 
+ */
+function handleNicknameDisplay(responseJson) {
+  const nickname = responseJson['nickname'];
+  const createNicknameText = document.getElementById('createNickname'); 
+
+  if (nickname === undefined || nickname === null) {
```
jamesnam-google: I think in Javascript it might be equivalent to,

if (!nickname) {
  ...
}

But please double check before simplifying the logic this way.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,143 +1,155 @@
...
     response.setContentType("application/json");
-    response.getWriter().println(quotesJson); 
+    UserService userService = UserServiceFactory.getUserService();
+
+    if (userService.isUserLoggedIn()) {
+      String userEmail = userService.getCurrentUser().getEmail();
+      String nickname = getUserNickname(userService.getCurrentUser().getUserId()); 
+
+      // HashMap that contains user information: email address, nickname.
+      HashMap<String, String> userInfo = new HashMap<String, String>();
```
jamesnam-google: Similarly, we can use the interface as the type.

Map<...>

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,143 +1,155 @@
...
+   * Creates a new entity for each quote and puts each new quote into Datastore
+   * @param request: the HTTP request; 
+   *        userInfo: a HashMap that contains user information connected to each quote
    */ 
-  @Override
-  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {    
-    String quote = request.getParameter("quote");
-
-    // Only update the new quote if it is not an empty string. 
+  private void putQuoteIntoDatastore(HttpServletRequest request, HashMap<String, String> userInfo) {
```
jamesnam-google: Map instead of HashMap

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,143 +1,155 @@
...
-    return value;
-  }
+    // Get the parameter for number to display, with 5 as the default value.
+    int numQuoteToDisplay = getNumQuoteDisplayed(request, DEFAULT_NUM_QUOTES);
+   
+    // Add each quote to the quote list whose content will be written as 
+    // the response from the servlet. 
+    List<Quote> quoteList = new ArrayList<>();
+    quotes.asList(FetchOptions.Builder.withLimit(numQuoteToDisplay))
+          .forEach((quoteEntity) -> {
```
jamesnam-google: Indentation seems excessive here.

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,143 +1,155 @@
...
+      responseMap.put(NICKNAME, nickname); 
+
+      // Only display quotes if the user is logged in
+      String quotesJson = getQuoteListJson(request);
+      responseMap.put(QUOTE, quotesJson);
+    } else {
+      String urlToRedirectToForLogIn = "/";
+      String loginUrl = userService.createLoginURL(urlToRedirectToForLogIn);
+      responseMap.put(LOGGEDIN, "false");
+      responseMap.put(USEREMAIL, null);
```
rgcase: Do we need all these fields with `null` values or can we just skip them?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,143 +1,155 @@
...
-    // This adds buffer time for querying Datastore. 
-    try {
-      Thread.sleep(200);
-    } catch (Exception e) {
-      System.out.println(e);
+    UserService userService = UserServiceFactory.getUserService();
+    HashMap<String, String> responseMap = new HashMap<String, String>();
+
+    if (userService.isUserLoggedIn()) {
+      String userEmail = userService.getCurrentUser().getEmail();
```
rgcase: Maybe extract the User first, so you don't have to keep calling `userService.getCurrentUser()`? You don't really know how expensive that call is, so I think it would be better to only call it once.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,143 +1,155 @@
...
+      String urlToRedirectToForLogOut = "/";
```
rgcase: Looks like this is used in both cases. Can we just extract this String into a constant like:
```
private static final String INDEX_URL = "/";
```

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,143 +1,155 @@
...
+      String logoutUrl = userService.createLogoutURL(urlToRedirectToForLogOut);
+      String nickname = getUserNickname(userService.getCurrentUser().getUserId());  
+
+      responseMap.put(LOGGEDIN, "true");
```
rgcase: I think we can model this information a little better. Consider maybe defining your own `User` class that uses a subset of the fields from the one you get back from the `UserService`. Convert to your version right after reading it out of the `UserService` and then you can just convert the whole thing to JSON and send it down. You also won't need this `LOGGGED_IN` field, because just the existence of the `User` object in the response is enough for the client to know that it's logged in. What do you think?

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -154,4 +166,35 @@ private int getNumQuoteDisplayed(HttpServletRequest request, int defaultNumToDis
...
+    String nickname = (String) quoteEntity.getProperty(NICKNAME); 
+    Quote newQuote = new Quote(id, text, timestampMillis, userEmail, nickname);
+    return newQuote; 
+  } 
+
+  /**
+   * Returns the nickname of the user with id, or empty String if the user has not set a nickname.
+   */
+  private String getUserNickname(String id) {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
```
rgcase: You'll probably want to use Strong consistency here. It might be nice to wrap this in a helper method so you can just call `getDatastoreService()` wherever you need it.

rgcase (changes requested):

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/webapp/script.js
hollyyuqizheng: Done. In the commits that follow, I will have factored out all the JavaScript code for the map into a separate file, so the new commit might show that I don't have this function included anymore. I will make the change accordingly in the new JS file for the map, which will be pushed later with the map commits.

portfolio/src/main/webapp/script.js
hollyyuqizheng: Yes, the ! symbol works in this case.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: I see -- yes, I realized that I can just skip them.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: I ended up not creating a User class, because there are enough helper methods within this servlet to handle the fetching of email address and nicknames.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,143 +1,153 @@
...
-import java.util.ArrayList; 
+import java.util.ArrayList;
+import java.util.Map; 
+import java.util.HashMap; 
 
-/** Servlet that returns some example content. */
 @WebServlet("/data")
 public class DataServlet extends HttpServlet {
 
   private static final String QUOTE = "Quote";
```
jamesnam-google: Consider using comments to distinguish different kinds of constants.

ex:

// Response param names
...

// URL constants
...

jamesnam-google: Also consider updating the constant variable names so they are more clear.

ex:

String PARAM_NAME_LOGGED_IN = ...

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -154,4 +164,35 @@ private int getNumQuoteDisplayed(HttpServletRequest request, int defaultNumToDis
...
+    String text = (String) quoteEntity.getProperty(TEXT);
+    long timestampMillis = (long) quoteEntity.getProperty(TIMESTAMP);
+    String userEmail = (String) quoteEntity.getProperty(USER_EMAIL); 
+    String nickname = (String) quoteEntity.getProperty(NICKNAME); 
+    Quote newQuote = new Quote(id, text, timestampMillis, userEmail, nickname);
+    return newQuote; 
+  } 
+
+  /**
+   * Returns the nickname of the user with id, or empty String if the user has not set a nickname.
```
jamesnam-google: Consider clarifying in the method comment that the method can return null.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,143 +1,153 @@
...
-   *         was not specified by the client
-   */ 
-  private String getParameter(HttpServletRequest request, String name, String defaultValue) {
-    String value = request.getParameter(name);
-    if (value == null) {
-      return defaultValue;
-    }
-    return value;
-  }
+    // Get the parameter for number to display, with 5 as the default value.
```
jamesnam-google: We don't need "with 5 as the default value" because then this comment has to be in sync with the constant variable.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,143 +1,153 @@
...
       datastore.put(quoteEntity);
     }
-    
-    response.sendRedirect("/index.html");
   }
 
   /**
-   * Extracts the attributes of a quoteEntity and puts them into a new
-   * instance of Quote. 
+   * Fetches quotes from Datastore and put them into a list in JSON format.
```
jamesnam-google: put => puts

portfolio/src/main/java/com/google/sps/servlets/DeleteDataServlet.java
```diff
@@ -35,19 +40,38 @@
...
+            keyList.add(quoteEntity.getKey());
+          } 
         });
 
     datastore.delete(keyList);
   }
+
+  /**
+   * Sets up strong consistency for datastore service. 
+   * This Strong Consistency ensures that freshness is more important than availability
```
jamesnam-google: nit: Strong Consistency => strong consistency

portfolio/src/main/java/com/google/sps/servlets/NicknameServlet.java
```diff
@@ -0,0 +1,112 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/nickname")
+public class NicknameServlet extends HttpServlet {
+
+  /**
+   * Handles GET request. If the user is logged in, display a page for them to input
```
jamesnam-google: display => displays

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -154,4 +164,35 @@ private int getNumQuoteDisplayed(HttpServletRequest request, int defaultNumToDis
...
+    String userEmail = (String) quoteEntity.getProperty(USER_EMAIL); 
+    String nickname = (String) quoteEntity.getProperty(NICKNAME); 
+    Quote newQuote = new Quote(id, text, timestampMillis, userEmail, nickname);
+    return newQuote; 
+  } 
+
+  /**
+   * Returns the nickname of the user with id, or empty String if the user has not set a nickname.
+   */
+  private String getUserNickname(String id) {
```
jamesnam-google: Consider renaming id to userId, because there can be different kinds of IDs

portfolio/src/main/java/com/google/sps/servlets/NicknameServlet.java
```diff
@@ -0,0 +1,112 @@
...
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    UserService userService = UserServiceFactory.getUserService();
+    if (!userService.isUserLoggedIn()) {
+      response.sendRedirect("/nickname");
+      return;
+    }
+
+    String nickname = request.getParameter("nickname");
+    String id = userService.getCurrentUser().getUserId();
```
jamesnam-google: Similarly, consider renaming to userId

portfolio/src/main/java/com/google/sps/servlets/NicknameServlet.java
```diff
@@ -0,0 +1,112 @@
...
+
+    datastore.put(entity);
+
+    response.sendRedirect("/index.html");
+  }
+
+  /**
+   * Returns the nickname of the user with id, or empty String if the user has not set a nickname.
+   */
+  private String getUserNickname(String id) {
```
jamesnam-google: Same here about id

jamesnam-google: Earlier I noticed that you return null if the nickname is not found. I would suggest that this method do the same (return null rather than empty string).

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DeleteDataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/NicknameServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/NicknameServlet.java
hollyyuqizheng: Done

hollyyuqizheng: Done

portfolio/src/main/java/com/google/sps/servlets/NicknameServlet.java
hollyyuqizheng: Done

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done. I added some comments to group the constants together, but I didn't rename many of the variables.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,143 +1,159 @@
...
+
+    response.sendRedirect("/index.html");
+  } 
+
+  /**
+   * Sets up strong consistency for datastore service. 
+   * This strong consistency ensures that freshness is more important than availability
+   * so that the most up-to-date data is returned and displayed on the page.
+   */ 
+  private DatastoreService getDatastoreServiceWithConsistency() {
```
rgcase: I think you can make this method `static`.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,143 +1,159 @@
...
+   * Fetches quotes from Datastore and puts them into a list in JSON format. 
    */ 
-  private Quote extractQuoteFromEntity(Entity quoteEntity) {
-    long id = quoteEntity.getKey().getId();
-    String text = (String) quoteEntity.getProperty(TEXT);
-    long timestampMillis = (long) quoteEntity.getProperty(TIMESTAMP);
-    Quote newQuote = new Quote(id, text, timestampMillis);
-    return newQuote; 
-  }
+  private String getQuoteListJson(HttpServletRequest request) {
```
rgcase: It's generally a good practice to pass in the minimal set of arguments that you can, and I think in this case if we call `getNumQuoteDisplayed(...)` before this method then you can just pass in that count here instead of the whole request.

Passing in the minimal data makes it easier to reason about the method from the outside because there are just fewer things that the method could be doing, since it has access to less data.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -154,4 +170,35 @@ private int getNumQuoteDisplayed(HttpServletRequest request, int defaultNumToDis
...
+   */
+  private String getUserNickname(String userId) {
+    DatastoreService datastore = getDatastoreServiceWithConsistency();
+    Query query =
+        new Query("UserInfo")
+            .setFilter(new Query.FilterPredicate("userId", Query.FilterOperator.EQUAL, userId));
+    PreparedQuery results = datastore.prepare(query);
+    Entity entity = results.asSingleEntity();
+    if (entity == null) {
+      return null;
```
rgcase: This is a good use case to use `java.util.Optional` instead of returning `null`. `Optional` lets you represent a value that might not be there, similar to how `null` works but it's more explicit, so it signals to the caller (using the type system) that they should think about how to handle the empty case, whereas right now it's not obvious to the caller that this method might return null (even if it's in the javadoc).

`Optional` has methods like `.ifPresent(...)` that let you do something with the value inside, or you can explicitly check with `.isPresent()` and then call `.get()`.

You can create an `Optional` with `Optional.ofNullable(...)`.

portfolio/src/main/java/com/google/sps/servlets/DeleteDataServlet.java
```diff
@@ -35,19 +40,38 @@
...
 
     datastore.delete(keyList);
   }
+
+  /**
+   * Sets up strong consistency for datastore service. 
+   * This strong consistency ensures that freshness is more important than availability
+   * so that the most up-to-date data is returned and displayed on the page.
+   */ 
+  private DatastoreService getDatastoreServiceWithConsistency() {
```
rgcase: `static` here as well. You could even extract this method out into a Util class or something, and have both your servlets use the same method.

portfolio/src/main/java/com/google/sps/servlets/NicknameServlet.java
```diff
@@ -0,0 +1,112 @@
...
+    String nickname = (String) entity.getProperty("nickname");
+    return nickname;
+  }
+
+  /**
+   * Sets up strong consistency for datastore service. 
+   * This strong consistency ensures that freshness is more important than availability
+   * so that the most up-to-date data is returned and displayed on the page.
+   */ 
+  private DatastoreService getDatastoreServiceWithConsistency() {
```
rgcase: Same here.

portfolio/src/main/java/com/google/sps/servlets/NicknameServlet.java
```diff
@@ -0,0 +1,112 @@
...
+
+    datastore.put(entity);
+
+    response.sendRedirect("/index.html");
+  }
+
+  /**
+   * Returns the nickname of the user with id, or null if the user has not set a nickname.
+   */
+  private String getUserNickname(String userId) {
```
rgcase: This is used in a couple places so you could extract this as well out into a helper class so you don't have duplicate code in your servlets. Also the same comment about using `Optional`.

rgcase (changes requested):

portfolio/src/main/webapp/script.js
```diff
@@ -109,30 +109,90 @@ function getQuotes() {
...
-      const quoteListElement = document.getElementById('quoteWrapper');
-      quoteListElement.innerHTML = '';
-
-      Object.values(quotesJson).forEach(
-        (quote) => {
-          quoteListElement.appendChild(createQuotesListElement(quote)); 
-        });        
+    .then((responseJson) => {
+      updateLoginStatusElements(responseJson);
+      if (responseJson['Quote'] !== undefined) {
```
jamesnam-google: I think you might be just able to remove the !== undefined part.

if (responseJson['Quote']) {
  ...
}

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,143 +1,159 @@
...
+  private static final String NICKNAME = "nickname";
 
-  /**
-    * Retrives all quotes that are stored in Datastore
-    * and puts them into a list of quotes. This list is then written
-    * as the response from the server. 
-    */
+  // Constants for information to be put into response of a GET request.
+  private static final String IS_USER_LOGGED_IN = "loggedIn"; 
+  private static final String REDIRECT_URL = "redirectUrl";
```
jamesnam-google: I'd recommend renaming this to REDIRECT_URL_PARAM, because it doesn't seem like it's the URL string itself

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,143 +1,159 @@
...
     response.setContentType("application/json");
-    response.getWriter().println(quotesJson); 
+    UserService userService = UserServiceFactory.getUserService();
+
+    if (userService.isUserLoggedIn()) {
+      User currentUser = userService.getCurrentUser(); 
+      putQuoteIntoDatastore(request, currentUser);
+    } 
+
+    response.sendRedirect("/index.html");
```
jamesnam-google: Maybe we can make "/index.html" a constant as well.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -154,4 +170,35 @@ private int getNumQuoteDisplayed(HttpServletRequest request, int defaultNumToDis
...
+    String text = (String) quoteEntity.getProperty(TEXT);
+    long timestampMillis = (long) quoteEntity.getProperty(TIMESTAMP);
+    String userEmail = (String) quoteEntity.getProperty(USER_EMAIL); 
+    String nickname = (String) quoteEntity.getProperty(NICKNAME); 
+    Quote newQuote = new Quote(quoteId, text, timestampMillis, userEmail, nickname);
+    return newQuote; 
+  } 
+
+  /**
+   * Returns the nickname of the user with id, or null if the user has not set a nickname.
```
jamesnam-google: Consider clarifying in the method comment that it can return null.

portfolio/src/main/java/com/google/sps/servlets/NicknameServlet.java
```diff
@@ -0,0 +1,112 @@
...
+
+  /**
+   * Handles POST request by putting a nickname into Datastore and 
+   * associate it with a user id. 
+   */ 
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    UserService userService = UserServiceFactory.getUserService();
+    if (!userService.isUserLoggedIn()) {
+      response.sendRedirect("/nickname");
```
jamesnam-google: Consider making "/nickname" a constant variable.

portfolio/src/main/java/com/google/sps/servlets/NicknameServlet.java
```diff
@@ -0,0 +1,112 @@
...
+    entity.setProperty("userId", userId);
+    entity.setProperty("nickname", nickname);
+
+    datastore.put(entity);
+
+    response.sendRedirect("/index.html");
+  }
+
+  /**
+   * Returns the nickname of the user with id, or null if the user has not set a nickname.
```
jamesnam-google: Consider clarifying in the method comment that it can return null.

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: I moved this method into a helper class "Util" and changed the method to static.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: I see!

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: I incorporated Optional in this method. In other methods in other classes where I need to fetch a nickname, I also use an Optional and check if it holds a value.

portfolio/src/main/java/com/google/sps/servlets/DeleteDataServlet.java
hollyyuqizheng: Moved to helper class Util.

portfolio/src/main/java/com/google/sps/servlets/NicknameServlet.java
hollyyuqizheng: Moved to helper class Util.

portfolio/src/main/java/com/google/sps/servlets/NicknameServlet.java
hollyyuqizheng: Moved to helper class Util.

portfolio/src/main/webapp/script.js
hollyyuqizheng: I see. I also changed a few other parts where I was originally comparing variables to null or undefined.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: I incorporated Optional in this method.

portfolio/src/main/java/com/google/sps/servlets/NicknameServlet.java
hollyyuqizheng: Done.

portfolio/src/main/java/com/google/sps/servlets/NicknameServlet.java
hollyyuqizheng: I incorporated Optional in this method.

portfolio/src/main/webapp/style.css
```diff
@@ -105,7 +111,7 @@
   color: white;
```
jamesnam-google, line 112: Let's have these lines sorted.

Reference:
https://google.github.io/styleguide/htmlcssguide.html#Declaration_Order

jamesnam-google: Here and elsewhere.

portfolio/src/main/webapp/script.js
```diff
@@ -109,30 +109,91 @@ function getQuotes() {
...
+
+/**
+ * Creates a li element that contains each individual quote;
  * this li element will be added to the Quotes section in the HTML file.  
  */
 function createQuotesListElement(quote) {
   const liElement = document.createElement('li');
-  liElement.innerText = quote.text;
+  const nickname = quote.nickname;  
+  console.log(nickname);
```
jamesnam-google: Should we remove this line?

portfolio/src/main/webapp/script.js
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+
```
jamesnam-google: Unnecessary line break

portfolio/src/main/java/com/google/sps/servlets/DeleteDataServlet.java
```diff
@@ -14,38 +14,62 @@
...
 import java.util.List; 
 import java.util.ArrayList; 
+import java.util.Optional;
 
 /** Servlet responsible for deleting all quotes in Datastore. */
 @WebServlet("/deleteQuote")
 public class DeleteDataServlet extends HttpServlet {
 
   private static final String QUOTE = "Quote";
+  private static final String USER_EMAIL = "userEmail";
```
jamesnam-google: Consider renaming this to,

PROPERTY_NAME_USER_EMAIL

portfolio/src/main/java/com/google/sps/servlets/NicknameServlet.java
```diff
@@ -0,0 +1,101 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.util.Optional; 
+
+@WebServlet("/nickname")
+public class NicknameServlet extends HttpServlet {
+
+  private static final String NICKNAME = "nickname";
```
jamesnam-google: Consider renaming this to

PROPERTY_NAME_NICKNAME

portfolio/src/main/java/com/google/sps/servlets/NicknameServlet.java
```diff
@@ -0,0 +1,101 @@
...
+  private static final String USER_ID = "userId";
```
jamesnam-google: Same here. If we name this USER_ID, it gives the impression that it stores the value of the user ID.

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/servlets/NicknameServlet.java
hollyyuqizheng: I see, yes that makes sense.

portfolio/src/main/webapp/script.js
hollyyuqizheng: Yes, sorry I ignored it.

portfolio/src/main/webapp/script.js
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/servlets/DeleteDataServlet.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/servlets/NicknameServlet.java
hollyyuqizheng: Updated.

portfolio/src/main/webapp/style.css
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,143 +1,155 @@
...
+import java.util.Optional;
 
-/** Servlet that returns some example content. */
 @WebServlet("/data")
 public class DataServlet extends HttpServlet {
 
-  private static final String QUOTE = "Quote";
-  private static final String TEXT = "text";
-  private static final String TIMESTAMP = "timestamp"; 
+  private static final String INDEX_URL = "/index.html";
```
jamesnam-google, line 30: Consider placing this next to other similar constants like HOME_URL

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,143 +1,155 @@
...
+
+  // Constants for property of a Quote item in Datastore.
+  private static final String PROPERTY_NAME_QUOTE = "Quote";
+  private static final String PROPERTY_NAME_TEXT = "text";
+  private static final String PROPERTY_NAME_TIMESTAMP = "timestamp"; 
+  private static final String PROPERTY_NAME_USER_EMAIL = "userEmail";
+  private static final String PROPERTY_NAME_NICKNAME = "nickname";
+
+  // Constants for information to be put into response of a GET request.
+  private static final String IS_USER_LOGGED_IN = "loggedIn";
```
jamesnam-google: Consider renaming this to,

IS_USER_LOGGED_IN_PARAM

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,143 +1,155 @@
...
   }
 
   /**
-   * Handles the returning of request parameter. 
-   * @param request: the request from the survey
-            name: the request parameter, which corresponds to a specific entry of the survey
-            defaultValue: a default value for this survey entry's input
-   * @return the request parameter, or the default value if the parameter
-   *         was not specified by the client
+   * Creates a new entity for each quote and puts each new quote into Datastore
```
jamesnam-google: nit: Period at the end.

portfolio/src/main/webapp/style.css
```diff
@@ -143,12 +149,12 @@
 
 #projectWrapper {
   display: grid;
-  grid-template-columns: auto auto ; 
   grid-column-gap: 3%;
-  grid-row-gap: 0; 
-  margin-right: 20%; 
-  padding-top: 2%;
+  grid-row-gap: 0;
+  grid-template-columns: auto auto ;
```
jamesnam-google, line 154: nit: Unnecessary space before ;

portfolio/src/main/webapp/script.js
```diff
@@ -143,10 +202,14 @@ function cleanData() {
...
 /**
- * This function handles the visual refreshing of the quote list element 
+ * Handles the visual refreshing of the quote list element 
  * by setting its content to empty. 
  */
 function refreshQuotes() {
   const quoteListElement = document.getElementById('quoteWrapper');
   quoteListElement.innerHTML = '';
+
+  // Re-fetching all current quotes from Datastrore after deleting visually.
```
jamesnam-google: nit: Re-fetching => Refetch

jamesnam-google (approved):
Approving the changes with comments.

portfolio/src/main/java/com/google/sps/data/Util.java
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * @return an Optional of type String that represents the user's nickname. 
+   */
+  public Optional<String> getUserNickname(DatastoreService datastore, String userId) {
+    Query query =
+        new Query(PROPERTY_NAME_USER_INFO)
+            .setFilter(new Query.FilterPredicate(PROPERTY_NAME_USER_ID, Query.FilterOperator.EQUAL, userId));
+    PreparedQuery results = datastore.prepare(query);
+    Entity entity = results.asSingleEntity();
```
rgcase, line 41: I think a simpler way to do this could just be:
```
return Optional.ofNullable(results.asSingleEntity()).map(entity.getProperty(PROPERTY_NAME_NICKNAME));
```

portfolio/src/main/java/com/google/sps/data/Util.java
```diff
@@ -0,0 +1,51 @@
...
+    DatastoreServiceConfig datastoreConfig = 
+        DatastoreServiceConfig.Builder.withReadPolicy(new ReadPolicy(Consistency.STRONG)).deadline(5.0);
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService(datastoreConfig);
+    return datastore; 
+  }
+
+  /**
+   * @return an Optional of type String that represents the user's nickname. 
+   */
+  public Optional<String> getUserNickname(DatastoreService datastore, String userId) {
```
rgcase: You can make this `static` as well.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,143 +1,155 @@
...
+  // Constant for fetching quotes from Datastore
   private static final int DEFAULT_NUM_QUOTES = 5;
 
-  /**
-    * Retrives all quotes that are stored in Datastore
-    * and puts them into a list of quotes. This list is then written
-    * as the response from the server. 
-    */
+  // Global variable for an instance of the Util helper class
+  private static Util util;
```
rgcase: Once you make the `getUserNickname(...)` method static you don't need an instance of `Util` here because you can just call the static methods directly like `Util.getDatastoreServiceWithConsistency()`.

rgcase (approved):

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
hollyyuqizheng: Updated.

portfolio/src/main/webapp/style.css
hollyyuqizheng: Updated.

portfolio/src/main/webapp/script.js
hollyyuqizheng: Updated.

PR #6, opened 2020-06-16
Week 5

hollyyuqizheng:
[Link](https://github.com/googleinterns/step/blob/master/walkthroughs/week-5-tdd/project/calendar-walkthrough.md)

This Pull Request contains both parts of the week 5 project: scheduling for mandatory attendees as well as for optional attendees. As the instructions required, the algorithm will pick time slots if one or more time slots exists so that both mandatory and optional attendees can attend. Otherwise, the algorithm returns the time slots that fit just the mandatory attendees.

I added the additional 5 tests for optional attendees as the instructions required, as well as a few more tests I created on my own. All tests pass.


walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,205 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.ArrayList;
 
 public final class FindMeetingQuery {
+
+  // Constants for 0 and the number of minutes in a day. 
+  // 0 means 0 minute since start of a day. 
+  // 1440 is the total number of minutes in a day, which marks the end of a day. 
+  private static final int START_OF_DAY = 0;
```
jamesnam-google: Consider renaming the variable so it indicates the time unit.

ex:
START_OF_DAY_MINUTES

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,205 @@
...
+import java.util.List;
+import java.util.ArrayList;
 
 public final class FindMeetingQuery {
+
+  // Constants for 0 and the number of minutes in a day. 
+  // 0 means 0 minute since start of a day. 
+  // 1440 is the total number of minutes in a day, which marks the end of a day. 
+  private static final int START_OF_DAY = 0; 
+  private static final int END_OF_DAY = 1440;
```
jamesnam-google: Same here.

END_OF_DAY_MINUTES

Also, in this case we can perform the computation so it's clear how this value came about.

ex:

END_OF_DAY_MINUTES = 24 * 60;

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,205 @@
...
+    List<TimeRange> mandatoryAttendeesBusyTimeslots = getAllBusyTimeslots(events, mandatoryAttendees); 
+    List<TimeRange> optionalAttendeesBusyTimeslots = getAllBusyTimeslots(events, optionalAttendees);
+
+    sortBusyTimeslotsByStart(mandatoryAttendeesBusyTimeslots); 
+    sortBusyTimeslotsByStart(optionalAttendeesBusyTimeslots);
+
+    List<TimeRange> mandatoryAttendeesEmptySlots = getEmptySlots(mandatoryAttendeesBusyTimeslots, meetingDuration); 
+    List<TimeRange> optionalAttendeesEmptySlots = getEmptySlots(optionalAttendeesBusyTimeslots, meetingDuration);
+
+    if (mandatoryAttendees.size() == 0) {
```
jamesnam-google: When you're checking emptiness we can call isEmpty()

if (mandatoryAttendees.isEmpty()) {
  ...

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,205 @@
...
+      return optionalAttendeesEmptySlots; 
+    } else if (optionalAttendees.size() == 0) {
```
jamesnam-google: Same here

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,205 @@
...
+        busyTimes.remove(busyTimeRange);
+        busyTimes.add(combinedTimeRange);
+        overlapped = true;  
+        currEventTimeRange = combinedTimeRange; 
+      } 
+    }
+
+    // If the current event doesn't overlap with any existing busy time range,
+    // add the current event's time range to the set of all busy time slots. 
+    if (! overlapped) {
```
jamesnam-google: We don't need the space after !

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,205 @@
...
+      busyTimes.add(currEventTimeRange); 
+    }
+  }
+
+  /** Creates a new overall combined time range for two time ranges that overlap */
```
jamesnam-google: Can we have a period at the end of the sentence?

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,205 @@
...
+    int aStart = a.start();
+    int aEnd = a.end();
+    int bStart = b.start();
+    int bEnd = b.end();
+
+    TimeRange combinedTimeRange = TimeRange.fromStartEnd(
+        Math.min(aStart, bStart), 
+        Math.max(aEnd, bEnd),
+        false 
+    );
```
jamesnam-google: The closing paren can go to the previous line (after false), like false);

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,205 @@
...
+  private TimeRange getCombinedTimeRange(TimeRange a, TimeRange b) {
+    int aStart = a.start();
+    int aEnd = a.end();
+    int bStart = b.start();
+    int bEnd = b.end();
+
+    TimeRange combinedTimeRange = TimeRange.fromStartEnd(
+        Math.min(aStart, bStart), 
+        Math.max(aEnd, bEnd),
+        false
```
jamesnam-google: Can we have param comment for false?

/* <param_name>= */ false

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,205 @@
...
+    TimeRange combinedTimeRange = TimeRange.fromStartEnd(
+        Math.min(aStart, bStart), 
+        Math.max(aEnd, bEnd),
+        false 
+    ); 
+
+    return combinedTimeRange; 
+  }
+
+  /** Sort a list of busy time ranges by their start time. */
```
jamesnam-google: Sort => Sorts

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,205 @@
...
+    List<TimeRange> emptySlots = new ArrayList<TimeRange>();
+
+    if (allBusyTimeslots.size() == 0 && meetingDuration <= END_OF_DAY) {
+      emptySlots.add(TimeRange.fromStartEnd(0, END_OF_DAY, false));
+      return emptySlots; 
+    }
+
+    // A for loop is necessary because each consecutive pair 
+    // of time ranges needs to be looked at at each iteration. 
+    for (int i = 0; i < allBusyTimeslots.size(); i++ ) {
```
jamesnam-google: We don't need the space after i++

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,205 @@
...
+
+      if (i == 0) {
+        int earliestStart = allBusyTimeslots.get(i).start();
+        if (earliestStart >= meetingDuration) {
+          emptySlots.add(TimeRange.fromStartEnd(START_OF_DAY, earliestStart, false)); 
+        }
+      } 
+
+      if (i == allBusyTimeslots.size()-1) {
```
jamesnam-google: Spaces around -

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,205 @@
...
+        int latestEnd = allBusyTimeslots.get(i).end();
+        if ((END_OF_DAY - latestEnd) >= meetingDuration) {
+          emptySlots.add(TimeRange.fromStartEnd(latestEnd, END_OF_DAY, false)); 
+        }
+      } else {
+        int currEnd = allBusyTimeslots.get(i).end();
+        int nextStart = allBusyTimeslots.get(i+1).start();
```
jamesnam-google: Spaces around +

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
jamesnam-google: Here and below.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,205 @@
...
+          emptySlots.add(TimeRange.fromStartEnd(currEnd, nextStart, false)); 
+        }
+      }
+    } 
+    return emptySlots; 
+  }
+
+  /** Finds the common time range between time ranges of two collections */ 
+  private List<TimeRange> findCommonEmptySlots(
+      List<TimeRange> mandatoryAttendeesEmptySlots, List<TimeRange> optionalAttendeesEmptySlots, long meetingDuration) {
```
jamesnam-google: This line is kind of long. Consider using line breaks after commas

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,205 @@
...
+          } else {
+            commonSlots.add(mandatorySlot); 
+          }
+        } 
+
+        // Move the pointer pointing at the time range that ends earlier,
+        // because the time range that ends later might have intersection with other
+        // time ranges in the time ranges taht come after the one that ends earlier.
+        if (mandatorySlot.end() >= optionalSlot.end()) {
+          j ++;
```
jamesnam-google: We don't need the space before ++. Here and below.

walkthroughs/week-5-tdd/intro/src/main/java/com/google/sps/Greeter.java
```diff
@@ -22,6 +22,7 @@
    * Returns a greeting for the given name.
    */
   public String greet(String name) {
-    return "Hello " + name;
+
+    return "Hello " + name.trim();
```
jamesnam-google, line 25: FYI, it can be simpler to use String.format if there are multiple arguments.

ex
return String.format("Hello %s", name.trim());

But concatenation works too.

walkthroughs/week-5-tdd/intro/src/test/java/com/google/sps/GreeterTest.java
```diff
@@ -30,4 +30,14 @@ public void testGreeting() {
...
     Assert.assertEquals("Hello Ada", greeting);
   }
+
+  @Test
+  public void testGreetingTrimsWhitespace() {
+    Greeter greeter = new Greeter();
+
+    String greeting = greeter.greet("   Ada   ");
+
+    // Whitespace should be trimmed
```
jamesnam-google: nit: Consider putting a period at the end.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,205 @@
...
+  private static final int END_OF_DAY = 1440; 
+
   public Collection<TimeRange> query(Collection<Event> events, MeetingRequest request) {
-    throw new UnsupportedOperationException("TODO: Implement this method.");
+
+    Collection<String> mandatoryAttendees = request.getAttendees();
+    Collection<String> optionalAttendees = request.getOptionalAttendees(); 
+    long meetingDuration = request.getDuration(); 
+
+    List<TimeRange> mandatoryAttendeesBusyTimeslots = getAllBusyTimeslots(events, mandatoryAttendees);
```
jamesnam-google: I noticed that sometimes we use the term Timeslots, sometimes we omit time (like EmptySlots), etc. I would suggest that we use consistent terminology.

For example, we can consistently write TimeSlots or TimeRanges.

ex:
mandatoryAttendeesEmptyTimeSlots
getAllBusyTimeSlots

Note the capitalized S in Slots.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,205 @@
...
+
+    events.forEach((event) -> {
+      Collection<String> eventAttendees = event.getAttendees();
+
+      for (String attendee : eventAttendees) {
+        // If the current event's attendee list contains a mandatory attendee, 
+        // adds this event's time range to the set of all busy times, 
+        // after checking for overlapping time ranges. 
+        if (mandatoryAttendees.contains(attendee)) {
+          TimeRange currEventTimeRange = event.getWhen();
```
jamesnam-google: nit: I usually try not to abridge words, like curr, and just write current. But it's up to you.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,205 @@
...
+      return allEmptySlots;
+    }
+  }
+
+  /** 
+   * Constructs all the time ranges that any of the mandatory attendees is busy.
+   * These time ranges cannot be part of the potential meeting time. 
+   */ 
+  private List<TimeRange> getAllBusyTimeslots(Collection<Event> events, Collection<String> mandatoryAttendees) {
+    List<TimeRange> busyTimes = new ArrayList<TimeRange>();
```
jamesnam-google: Same comment here about terminology around time slots. Maybe you can rename it to busyTimeSlots. I would personally prefer busyTimeRanges, because the type is TimeRange. But TimeSlot works too.

jamesnam-google (changes requested):

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/intro/src/main/java/com/google/sps/Greeter.java
hollyyuqizheng: Got it!

walkthroughs/week-5-tdd/intro/src/test/java/com/google/sps/GreeterTest.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Got it!

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,220 @@
...
+    Collection<String> optionalAttendees = request.getOptionalAttendees(); 
+    long meetingDuration = request.getDuration(); 
+
+    List<TimeRange> mandatoryAttendeesBusyTimeRanges = getAllBusyTimeRanges(events, mandatoryAttendees); 
+    List<TimeRange> optionalAttendeesBusyTimeRanges = getAllBusyTimeRanges(events, optionalAttendees);
+
+    sortBusyTimeRangesByStart(mandatoryAttendeesBusyTimeRanges); 
+    sortBusyTimeRangesByStart(optionalAttendeesBusyTimeRanges);
+
+    List<TimeRange> mandatoryAttendeesEmptyRanges = getEmptyTimeRanges(mandatoryAttendeesBusyTimeRanges, meetingDuration);
```
jamesnam-google: This line and the one below seem kind of long. Can we use line breaks?

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,220 @@
...
 
 public final class FindMeetingQuery {
+
+  // Constants for 0 and the number of minutes in a day. 
+  // 0 means 0 minute since start of a day. 
+  // 1440 is the total number of minutes in a day, which marks the end of a day. 
+  private static final int START_OF_DAY_MINUTES = 0; 
+  private static final int END_OF_DAY_MINUTES = 24 * 60; 
+
   public Collection<TimeRange> query(Collection<Event> events, MeetingRequest request) {
```
jamesnam-google: Consider adding comment to the public method.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,220 @@
...
+    TimeRange combinedTimeRange = TimeRange.fromStartEnd(
+        /* start time = */ Math.min(aStart, bStart), 
+        /* end time = */ Math.max(aEnd, bEnd),
+        /* inclusive of end time = */ false); 
+
+    return combinedTimeRange; 
+  }
+
+  /** Sorts a list of busy time ranges by their start time. */
+  private void sortBusyTimeRangesByStart(List<TimeRange> allBusyTimeRanges) {
```
jamesnam-google: I think this method does not necessarily only apply to busy time ranges. So consider removing the busy qualifier.

ex:
sortTimeRangesByStart(List<TimeRange> timeRanges)

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,220 @@
...
+  /**
+   * Finds all empty time ranges based on a list of busy time ranges and 
+   * the length of the requested meeting
+   */ 
+  private List<TimeRange> getEmptyTimeRanges(List<TimeRange> allBusyTimeRanges, long meetingDuration) {
+    List<TimeRange> emptyTimeRanges = new ArrayList<TimeRange>();
+
+    if (allBusyTimeRanges.size() == 0 && meetingDuration <= END_OF_DAY_MINUTES) {
+      emptyTimeRanges.add(TimeRange.fromStartEnd(
+          /* start time = */ 0,
```
jamesnam-google: The param comments need to match the param names.

ex:
/* start= */ ...
/* end= */ ...
/* inclusive= */ ...

jamesnam-google: Here and below.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,220 @@
...
+  }
+
+  /**
+   * Finds all empty time ranges based on a list of busy time ranges and 
+   * the length of the requested meeting
+   */ 
+  private List<TimeRange> getEmptyTimeRanges(List<TimeRange> allBusyTimeRanges, long meetingDuration) {
+    List<TimeRange> emptyTimeRanges = new ArrayList<TimeRange>();
+
+    if (allBusyTimeRanges.size() == 0 && meetingDuration <= END_OF_DAY_MINUTES) {
```
jamesnam-google: isEmpty() instead of size() == 0

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,220 @@
...
+  /** Sorts a list of busy time ranges by their start time. */
+  private void sortBusyTimeRangesByStart(List<TimeRange> allBusyTimeRanges) {
+    Collections.sort(allBusyTimeRanges, TimeRange.ORDER_BY_START); 
+  }
+
+  /**
+   * Finds all empty time ranges based on a list of busy time ranges and 
+   * the length of the requested meeting
+   */ 
+  private List<TimeRange> getEmptyTimeRanges(List<TimeRange> allBusyTimeRanges, long meetingDuration) {
```
jamesnam-google: Can we clarify the time unit for meetingDuration?

ex:
meetingDurationMinutes

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,220 @@
...
+    List<TimeRange> emptyTimeRanges = new ArrayList<TimeRange>();
+
+    if (allBusyTimeRanges.size() == 0 && meetingDuration <= END_OF_DAY_MINUTES) {
+      emptyTimeRanges.add(TimeRange.fromStartEnd(
+          /* start time = */ 0, 
+          /* end time = */ END_OF_DAY_MINUTES, 
+          /* inclusive of end time = */ false));
```
jamesnam-google: What if the meeting duration is exactly END_OF_DAY_MINUTES?

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,220 @@
...
+      }
+    } 
+    return emptyTimeRanges; 
+  }
+
+  /** Finds the common time range between time ranges of two collections */ 
+  private List<TimeRange> findCommonEmptyTimeRanges(
+      List<TimeRange> mandatoryAttendeesEmptyTimeRanges, 
+      List<TimeRange> optionalAttendeesEmptyTimeRanges, 
+      long meetingDuration) {
```
jamesnam-google: Similarly, can we clarify the time unit?

walkthroughs/week-5-tdd/project/src/test/java/com/google/sps/FindMeetingQueryTest.java
```diff
@@ -270,5 +276,181 @@ public void notEnoughRoom() {
...
+            Arrays.asList(PERSON_C)));
+ 
+    MeetingRequest request =
+        new MeetingRequest(Arrays.asList(PERSON_A, PERSON_B), DURATION_30_MINUTES);
+      
+    request.addOptionalAttendee(PERSON_C); 
+ 
+    Collection<TimeRange> actual = query.query(events, request);
+    Collection<TimeRange> expected =
+        Arrays.asList(TimeRange.fromStartEnd(TimeRange.START_OF_DAY, TIME_0800AM, false),
```
jamesnam-google: Param comment for false/true. Here and below.

jamesnam-google: We can have a line break before TimeRange.

ex:
Arrays.asList(
    TimeRange....,
    TimeRange...,
    TimeRange...);

jamesnam-google: (For some reason GitHub does not show the spaces)

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
jamesnam-google: Also, I wonder whether empty is the most descriptive term we can use. What do you think about free (which is conceptually the opposite of busy) ?

ex:
mandatoryAttendeesFreeTimeRanges

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
jamesnam-google: Similar comment about empty vs free

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,220 @@
...
+
+    // If the current event doesn't overlap with any existing busy time range,
+    // add the current event's time range to the set of all busy time slots. 
+    if (!overlapped) {
+      busyTimes.add(currEventTimeRange); 
+    }
+  }
+
+  /** Creates a new overall combined time range for two time ranges that overlap. */ 
+  private TimeRange getCombinedTimeRange(TimeRange a, TimeRange b) {
```
jamesnam-google: Consider making this method static.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,220 @@
...
+   * @param current event's time range; a collection of busy time slots
+   */ 
+  private void handleOverlapTimeRange(TimeRange currEventTimeRange, Collection<TimeRange> busyTimes) {
+    boolean overlapped = false; 
+
+    for (TimeRange busyTimeRange : busyTimes) {
+      if (currEventTimeRange.overlaps(busyTimeRange)) {
+        TimeRange combinedTimeRange = getCombinedTimeRange(currEventTimeRange, busyTimeRange);
+        busyTimes.remove(busyTimeRange);
+        busyTimes.add(combinedTimeRange);
```
jamesnam-google: Here you are making changes to busyTimes as you are iterating over busyTimes. Instead, consider updating handleOverlapTimeRange so it returns a new collection of busy time ranges, without modifying the given busyTimes. Then perhaps we can rename handleOverlapTimeRange to,

Collection<TimeRange> getUpdatedBusyTimeRanges

jamesnam-google (changes requested):

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: I see. The variable name is updated to "FreeTimeRanges" now.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: I noticed that the TimeRange class already has START_OF_DAY and END_OF_DAY defined as constants, and in that class, end of day is defined as 1439, which is the number of minutes for 23:59. I changed my code so that these two constants from TimeRange are used instead of my own defined constants for 0 and 1440. Based on TimeRange class's definition of the end of the day, I changed the "inclusive" parameter to true whenever the end time is TimeRaneg.END_OF_DAY. I also added a test to confirm my code change works. In my test, a 24-hour meeting is defined as having 1439 minutes.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,235 @@
...
+   * Picks time slots if one or more time slots 
+   * exists so that both mandatory and optional attendees can attend. 
+   * Otherwise, returns the time slots that fit just the mandatory attendees.
+   */ 
   public Collection<TimeRange> query(Collection<Event> events, MeetingRequest request) {
-    throw new UnsupportedOperationException("TODO: Implement this method.");
+
+    Collection<String> mandatoryAttendees = request.getAttendees();
+    Collection<String> optionalAttendees = request.getOptionalAttendees(); 
+    long meetingDuration = request.getDuration();
```
jamesnam-google: Consider indicating the time unit. ex: meetingDurationMinutes

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,235 @@
...
+        getFreeTimeRanges(mandatoryAttendeesBusyTimeRanges, meetingDuration); 
+    List<TimeRange> optionalAttendeesFreeTimeRanges = 
+        getFreeTimeRanges(optionalAttendeesBusyTimeRanges, meetingDuration);
+
+    if (mandatoryAttendees.isEmpty()) {
+      return optionalAttendeesFreeTimeRanges; 
+    } else if (optionalAttendees.isEmpty()) {
+      return mandatoryAttendeesFreeTimeRanges; 
+    } else {
+      List<TimeRange> allFreeTimeRanges = findCommonFreeTimeRanges(
```
jamesnam-google: (optional) Here you can just return findCommon... instead of declaring the variable.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,235 @@
...
+
+  /**
+   * Compares a given event's time slot with all the existing busy time slots.
+   * Merges the current event time slot with a busy time slot if they overlap.
+   * Adds the current event's time slot to the list of busy slots in the end. 
+   * @param current event's time range; a collection of busy time slots
+   * @return an updated list of busy times after overlaps are handled
+   */ 
+  private List<TimeRange> getUpdatedBusyTimeRanges(
+    TimeRange currEventTimeRange, Collection<TimeRange> originalBusyTimes) {
```
jamesnam-google: This line needs two more spaces of indentation.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,235 @@
...
+    boolean overlapped = false;
```
jamesnam-google: I usually name my boolean variables do... is... etc. So like doOverlap

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,235 @@
...
+            /* start= */ currEnd, 
+            /* end= */ nextStart, 
+            /* inclusive= */ false)); 
+        }
+      }
+    } 
+    return freeTimeRanges; 
+  }
+
+  /** Finds the common time range between time ranges of two collections */
```
jamesnam-google: nit: Period at the end.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,235 @@
...
+    return freeTimeRanges; 
+  }
+
+  /** Finds the common time range between time ranges of two collections */ 
+  private List<TimeRange> findCommonFreeTimeRanges(
+      List<TimeRange> mandatoryAttendeesFreeTimeRanges, 
+      List<TimeRange> optionalAttendeesFreeTimeRanges, 
+      long meetingDurationMinutes) {
+        
+      // If any of mandatory or optional attendee has no free time,
```
jamesnam-google: This line (and the lines below) are indented too much by two spaces.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,235 @@
...
+        TimeRange optionalTimeRange = optionalAttendeesFreeTimeRanges.get(j);
+
+        // If two time ranges overlap, find the intersection of these two. 
+        if (mandatoryTimeRange.overlaps(optionalTimeRange)) {
+          int start = Math.max(mandatoryTimeRange.start(), optionalTimeRange.start());
+          int end = Math.min(mandatoryTimeRange.end(), optionalTimeRange.end());
+          
+          if (end - start >= meetingDurationMinutes) {
+            TimeRange commonTimeRange = TimeRange.fromStartEnd(
+              /* start= */ start,
```
jamesnam-google: This line has to be indented two spaces further, as well as the lines below.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,235 @@
...
+
+    sortTimeRangesByStart(mandatoryAttendeesBusyTimeRanges); 
+    sortTimeRangesByStart(optionalAttendeesBusyTimeRanges);
+
+    List<TimeRange> mandatoryAttendeesFreeTimeRanges = 
+        getFreeTimeRanges(mandatoryAttendeesBusyTimeRanges, meetingDuration); 
+    List<TimeRange> optionalAttendeesFreeTimeRanges = 
+        getFreeTimeRanges(optionalAttendeesBusyTimeRanges, meetingDuration);
+
+    if (mandatoryAttendees.isEmpty()) {
```
jamesnam-google: Thanks for simplifying the logic here.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,235 @@
...
+   * Merges the current event time slot with a busy time slot if they overlap.
+   * Adds the current event's time slot to the list of busy slots in the end. 
+   * @param current event's time range; a collection of busy time slots
+   * @return an updated list of busy times after overlaps are handled
+   */ 
+  private List<TimeRange> getUpdatedBusyTimeRanges(
+    TimeRange currEventTimeRange, Collection<TimeRange> originalBusyTimes) {
+    boolean overlapped = false; 
+
+    List<TimeRange> updatedBusyTimes = new ArrayList<TimeRange> ();
```
jamesnam-google: nit: Unnecessary space before ()

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,235 @@
...
+    }
+
+    // A for loop is necessary because each consecutive pair 
+    // of time ranges needs to be looked at at each iteration. 
+    for (int i = 0; i < allBusyTimeRanges.size(); i++) {
+      if (i == 0) {
+        int earliestStart = allBusyTimeRanges.get(i).start();
+        if (earliestStart >= meetingDurationMinutes) {
+          freeTimeRanges.add(TimeRange.fromStartEnd(
+            /* start= */ TimeRange.START_OF_DAY,
```
jamesnam-google: This line has to be indented two spaces further. Here and below.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -15,9 +15,235 @@
...
+      // return the other group's free time slots. 
+      if (mandatoryAttendeesFreeTimeRanges.isEmpty()) {
+        return optionalAttendeesFreeTimeRanges;
+      }
+      if (optionalAttendeesFreeTimeRanges.isEmpty()) {
+        return mandatoryAttendeesFreeTimeRanges; 
+      }    
+
+      List<TimeRange> commonTimeRanges = new ArrayList<TimeRange>();
+      int i = 0;
```
jamesnam-google: Consider adding some comment that explains that i is for mandatory attendees and j is for optional attendees.

jamesnam-google (approved):
Approving the change with comments.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -14,10 +14,236 @@
...
+
+    sortTimeRangesByStart(mandatoryAttendeesBusyTimeRanges);
+    sortTimeRangesByStart(optionalAttendeesBusyTimeRanges);
+
+    List<TimeRange> mandatoryAttendeesFreeTimeRanges =
+        getFreeTimeRanges(mandatoryAttendeesBusyTimeRanges, meetingDurationMinutes);
+    List<TimeRange> optionalAttendeesFreeTimeRanges =
+        getFreeTimeRanges(optionalAttendeesBusyTimeRanges, meetingDurationMinutes);
+
+    if (mandatoryAttendees.isEmpty()) {
```
rgcase: I think you might be able to make this simpler. Right now you find the open timeslots for mandatory attendees, and open timeslots for optional attendees, then find the intersection between the two. But couldn't you instead call `getFreeTimeRanges(...)` with `mandatoryAttendees ++ optionalAttendees` to find the same intersection? Then you don't need the `findCommonFreeTimeRanges` method at all I think.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -14,10 +14,236 @@
...
+   * cannot be part of the potential meeting time.
+   */
+  private List<TimeRange> getAllBusyTimeRanges(
+      Collection<Event> events, Collection<String> mandatoryAttendees) {
+    List<TimeRange> busyTimes = new ArrayList<TimeRange>();
+
+    for (Event event : events) {
+      Collection<String> eventAttendees = event.getAttendees();
+
+      for (String attendee : eventAttendees) {
```
rgcase: I think you could make this a little cleaner and more efficient.

The first thing is that you're calling `.contains(...)` on the `Collection<String> mandatoryAttendees` which could be inefficient if the collection is large, so maybe convert it to a `HashSet` first which offers more efficient lookup.

Second, the logic for checking if the Event is interesting to us, and the logic for updating `busyTimes` is kind of mixed together when I think we could separate them a bit and reduce the level of nesting in the method. You could first check if any attendees are in our Set, and if not continue to the next Event, and if so then do the next piece of logic. Something like this:
```
Set<String> mandatoryAttendeeSet = new HashSet<>();
...
for (Event event : events) {
  boolean hasMandatoryAttendee =
    event.getAttendees().stream().filter(mandatoryAttendeeSet::contains).findAny().isPresent();

  if (!hasMandatoryAttendee) {
    continue;
  }

  ... build new busyTimes...
```

What do you think?

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -14,10 +14,236 @@
...
+    }
+  }
+
+  /**
+   * Constructs all the time ranges that any of the mandatory attendees is busy. These time ranges
+   * cannot be part of the potential meeting time.
+   */
+  private List<TimeRange> getAllBusyTimeRanges(
+      Collection<Event> events, Collection<String> mandatoryAttendees) {
+    List<TimeRange> busyTimes = new ArrayList<TimeRange>();
```
rgcase: Since you're going to update this `busyTimes` variable to an entirely new `List` on each iteration, this is a good opportunity to use `ImmutableList` from Guava if you like. You can find more about it (including how to add the dependency to your project) here https://github.com/google/guava.

You don't have to do it though for this if you don't want to, but if you're interested you can.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -14,10 +14,236 @@
...
+        }
+      }
+    }
+    return busyTimes;
+  }
+
+  /**
+   * Compares a given event's time slot with all the existing busy time slots. Merges the current
+   * event time slot with a busy time slot if they overlap. Adds the current event's time slot to
+   * the list of busy slots in the end.
```
rgcase: nit: The last sentence happens only when the current event's time slot doesn't overlap.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -14,10 +14,236 @@
...
+    int aStart = a.start();
+    int aEnd = a.end();
+    int bStart = b.start();
+    int bEnd = b.end();
+
+    TimeRange combinedTimeRange =
+        TimeRange.fromStartEnd(
+            /* start= */ Math.min(aStart, bStart),
+            /* end= */ Math.max(aEnd, bEnd),
+            /* inclusive= */ false);
```
rgcase: Doesn't this `inclusive` value have to depend on `a` and `b`? Like if say `bEnd > aEnd` and `b.inclusive() == true`, then you should pass `true` here I think?

rgcase (changes requested):

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: I see, yes this division of logic makes sense. I have updated the code accordingly.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: Updated.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: I think, based on the TimeRange class that is given to us, there isn't really a way to get the "inclusive" attribute of a TimeRange. From reading the class and the provided test cases, it seems that the default is exclusive for end time. The exception is if the end time is END_OF_DAY, then this time range should be inclusive of the end. 

In my method, I will change it so that, if the later end time between a and b is the END_OF_DAY, then I set the inclusive flag to true, otherwise the newly created time range is always exclusive of the end time.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: I wonder if busyTimes in this method is immutable? Based on my implementation right now, I update busyTimes to the output of `getUpdatedBusyTimeRanges(currEventTimeRange, busyTimes);` -- wouldn't this make busyTimes not immutable?

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
hollyyuqizheng: I tried changing the logic by calling `getFreeTimeRanges` on the combined attendee busy times, but some of the edge case tests with optional attendees won't pass. 

My thought is, if I combine the busy times of both mandatory and optional attendees and find the free times from there, there isn't a way for the `getFreeTimeRanges` method to know which busy time is from optional attendees and which busy time is from mandatory ones, at least not from how I have implemented that method. The overall algorithm is meant to find free time ranges for all optional and mandatory attendees, but if such time block doesn't exist, the algorithm should return all the blocks where the mandatory attendees can make it, so I feel that having free time ranges for mandatory and optional attendees separately is necessary here.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
rgcase: The variable `busyTimes` is mutable (not `final`) but the value that it points to is immutable because you always replace it with a completely new copy rather than updating the existing value in place. We can make a distinction between the mutability of the variable and the value that it points to:
```
List<String> foo = new ArrayList<>(); // Both variable and value mutable
final List<String> foo = new ArrayList<>(); // Variable immutable and value mutable
List<String> foo = ImmutableList.of(); // Variable mutable, value immutable
final List<String> foo = ImmutableList.of(); // Both immutable
```

In your case, I think you could define `ImmutableList<String> busyTimes` and keep updating the mutable variable with a new immutable list at each iteration.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
rgcase: I see ya you're right. Looking at the code in `TimeRange` they just subtract `1` from `end` if it's exclusive.

walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
rgcase: Ya you would need to compute `getFreeTimeRanges(...)` first for `mandatoryAttendees ++ optionalAttendees`, and if that's empty then do the same thing for just `mandatoryAttendees`. Did you try something like that?

hollyyuqizheng: I see -- I had ignored to combine the attendee list just now. I changed the code accordingly, this time getting all the busy and free time ranges for the combined attendee list first, and if this combined free time is empty, I run the same process on the mandatory attendee list.

rgcase (approved):

